PR #12, opened 2019-04-17
Logtext first implementation

shashikdm:
tensorboard frontend takes care of rendering markdown. We only need to log the String.
This is how it looks:
image
@PhilipVinc  Please elaborate on how you'd like to handle matrices and lists of Strings. Should matrix be converted to a table in markdown?


oxinabox:
Cool

The following should be of interest:
https://discourse.julialang.org/t/how-to-get-a-markdown-repr-of-an-object-mime-type-fallbacks/23242

Right now the best solution I have is at the bottom, and that solution is Ok.
not great, but Ok.  Might be the best that is possible with people writing poorly behaved `repr` implementations.

PhilipVinc:
Nice!

Does it also work if in the string you've got some Unicode symbols/emojis? 

Right now you only defined the function `text_summary(....text::String, ...)`. I think it would be a good idea to also define a more general function to convert arbitrary types to strings.  
```
function text_summary(name::String, text::Any)
      str = ... # convert any to markdown, maybe use Lyndon's trick ATM
      return text_summary(name, str)
end
```
Then you can also weaken the constraint of `log_text(logger, name, text::String; ...)` to `text::Any`. 

If there is no native way to do that in Julia, I believe that converting `Matrix{String}`  and `Vector{String}` to MD tables natively is a sane idea. It does look similar to the default REPR behaviour.

src/Loggers/LogText.jl
```diff
@@ -0,0 +1,41 @@
...
+    write_event(logger.file, make_event(logger, summ, step=step))
+end
+
+function text_summary(name::String, text::String)
+    texttensorshape = TensorShapeProto(dim = Vector([TensorShapeProto_Dim(size = 1)]))
+    textstringval = Vector{Array{UInt8, 1}}()
+    push!(textstringval, Array{UInt8, 1}(text))
+    pb = PipeBuffer()
+    writeproto(pb, TextPluginData(version = 0))
+    textcontent = pb.data
```
PhilipVinc: While you are at it, could you also define a method in `utils.jl`
```
function serialize_proto(data::Union{ProtoType, ProtoEnum})
     pb = PipeBuffer()
     writeproto(pb, data)
    pb.data
end
```

and use that? It will be used with practically every plugin and would make the code a bit easier to read.

src/utils.jl
```diff
@@ -5,3 +5,25 @@ function masked_crc32c(data)
...
+    pb = PipeBuffer()
+    writeproto(pb, data)
+    pb.data
+end
+
+"""
+    markdown_repr(x)
+
+Returns the most suitable method for displaying `x` in a markdown document.
+"""
```
oxinabox, line 19: See improved version in:
https://discourse.julialang.org/t/how-to-get-a-markdown-repr-of-an-object-mime-type-fallbacks/23242/3?u=oxinabox

PhilipVinc:
@shashikdm if you implement the "improved version"  I think this should be good to be merged.

PhilipVinc:
I forgot, could you also add a bunch of simple tests?

shashikdm:
Yes. I will add tests soon. My end semester exams will end on 3rd May, I will begin right after that

shashikdm:
Please note that codecov is failing because of unused Protobuf files. These are needed to compile Tensor.proto for `log_text` and will be actually used in future when `log_graph` will be implemented.

PhilipVinc:
Yes I noticed.
I will add a rule to ignore those two folders for coverage.

I’ll have a last look at this between today and tomorrow but I should merge this soon.
Il 27 apr 2019, 09:39 +0200, Shashikant Kadam <notifications@github.com>, ha scritto:

PhilipVinc:
Thanks a lot!

PR #21, opened 2019-05-08
Logimage

shashikdm:
started work on Logimage


shashikdm:
This is an inefficient initial implementation as I'm saving the file on disk in order to get a png file. Will change it to something like `IOBuffer()` soon.

Project.toml
```diff
@@ -6,6 +6,8 @@ version = "0.1.0"
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 CRC32c = "8bf52ea8-c179-5cab-976a-9e18b702a9bc"
+FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
+Images = "916415d5-f1e6-5110-898d-aaa5f9f070e0"
```
oxinabox: Images is a really large dependency as it is basically a metapackages for all of JuliaImages
If we just need `colorview` than I think `ImageCore` is fine

src/Loggers/LogImage.jl
```diff
@@ -1,14 +1,42 @@
...
-function image_summary(name::String, img)
-    Summary_Image(height = size(img)[1], width = size(img)[2], colorspace = 3, encoded_image_string = Nothing)
+function image_summary(name::AbstractString, img::Array)
+    #Encode image to string
+    io = IOBuffer()
+    iob64_encode = Base64.Base64EncodePipe(io)
+    write(iob64_encode, img)
+    close(iob64_encode)
+    str = String(take!(io));
+    eis = Vector{UInt8}(str)
```
PhilipVinc: Can't you do directly `eis=take!(io)` ?

src/Loggers/LogImage.jl
```diff
@@ -1,14 +1,42 @@
...
+    #obtain height and width of image
+    height = size(img)[1]
+    width = size(img)[2]
```
PhilipVinc:
```
height = size(img,1)
width = size(img,2)
```

src/Loggers/LogImage.jl
```diff
@@ -1,14 +1,42 @@
...
+
+
+"""
+import TensorBoardLogger
+using .TensorBoardLogger
+lgr = TBLogger("/home/shashi/run")
+img = rand(UInt8, 24, 24)
+log_value(lgr, "testval", 1226, step = 0)
+log_image(lgr, "testimage", img, step = 0)
+"""
```
PhilipVinc: Don't leave those python-codebase-style things there. 
But you could add a quick test using it.

src/Loggers/LogImage.jl
shashikdm: This is an old code. Please check out latest commit

src/Loggers/LogImage.jl
shashikdm: This is an old code. Please check out latest commit

src/Loggers/LogImage.jl
PhilipVinc: My bad. I'm on a train with shitty internet.

shashikdm:
TODO:
- Support `Image` objects along with Arrays
- Support multiple channels
- Take format as argument (`WHC`, `HWC`, `CHW` ...) and convert it to `HWC`
- Support `NHWC` as well
- AbstractLogger interface for log_image (`TBImage` Wrapper, `preprocess` function)
- testcases

PhilipVinc:
Leave aside the wrapper for now. I'm still thinking how to implement the wrapper logic, and it should just be 3 LOC...

shashikdm:
I will add tests tonight. please do not merge yet

shashikdm:
Screenshot_2019-05-12 TensorBoard(1)

shashikdm:
@PhilipVinc  please update Travic CI configs. tests are failing because I am unable to Use `Flux` and `Metalhead` packages in test file. I tried adding them using `Pkg.add("Flux")` but Error message says `Pkg` not found in current path 😅. It also says run `import Pkg; Pkg.add("Flux")` but where to write it?

test/runtests.jl
```diff
@@ -4,7 +4,8 @@ using Test
 @testset "TBLogger" begin
     include("test_TBLogger.jl")
 end
-
+Pkg.add("Flux")
+Pkg.add("Metalhead")
```
oxinabox: Install packages required for tests using the Project.toml.
https://docs.julialang.org/en/v1.0/stdlib/Pkg/#Test-specific-dependencies-1

Project.toml
```diff
@@ -12,6 +12,8 @@ ProtoBuf = "3349acd9-ac6a-5e09-bcdb-63829b23a429"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [extras]
+Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
+Metalhead = "dbeba491-748d-5e0e-a39e-b530a07fa0cc"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
```
shashikdm, line 18: I have tried this. But it won't install these packages. I get the same error. @oxinabox What am I doing wrong
image

oxinabox: You also need to added it below:
```
[targets]
test = ["Test", "Flux", "MetalHead"]
```

shashikdm:
@oxinabox @PhilipVinc please check if all requirements are met for `log_image`. Also, check if the code is efficient or not. Is there any dependency we could get rid of? I think `FileIO` is necessary to convert image to PNG format. `ImageMagick` should be installed while installing `FileIO`. but tests were failing so I added `ImageMagick` as a dependency too. `ImageCore` and `ColorTypes` are added for `[colorview, channelview]` and `[RGB, Gray, GrayA, RGBA]` respectively.

shashikdm:
I think it should also support 1-D Arrays. What do you think @PhilipVinc @oxinabox? Should it also support Array of images (`Array{Array{Float64, 3}, 1}`) or that can be handled in `preprocess` function?

oxinabox:
> @oxinabox @PhilipVinc please check if all requirements are met for log_image. Also, check if the code is efficient or not. Is there any dependency we could get rid of? I think FileIO is necessary to convert image to PNG format. ImageMagick should be installed while installing FileIO. but tests were failing so I added ImageMagick as a dependency to

This is normal, FileIO does not depend on ImageMagick (or anything else) by design.
You are supposed to depend on both.

PhilipVinc:
> I think it should also support 1-D Arrays. What do you think @PhilipVinc @oxinabox? Should it also support Array of images (`Array{Array{Float64, 3}, 1}`) or that can be handled in `preprocess` function?

`preprocess` is only relevant when using the Logger interface. For the explicit interface, it would be nice to have it but it's not essential for now.

shashikdm:
> > I think it should also support 1-D Arrays. What do you think @PhilipVinc @oxinabox? Should it also support Array of images (`Array{Array{Float64, 3}, 1}`) or that can be handled in `preprocess` function?
> 
> `preprocess` is only relevant when using the Logger interface. For the explicit interface, it would be nice to have it but it's not essential for now.

I can implement in two ways. overload `log_image` for `Array{Any, 1}` or new function `log_images` for `Array{Any, 1}`. I'm more inclined towards `log_images` because it makes its purpose clear. What's your opinion @oxinabox @PhilipVinc ?

oxinabox:
`log_images` seems better to me also.

src/Loggers/LogImage.jl
```diff
@@ -1,16 +1,35 @@
+
+@enum ImageFormat HW WH HWC WHC CHW CWH HWN WHN NHW NWH HWCN WHCN CHWN CWHN NHWC NWHC NCHW NCWH
```
PhilipVinc, line 2: I think it would be good to export ImageFormat.

shashikdm: i have exported `ImageFormat` in `TensorBoardLogger.jl`

PhilipVinc:
To me it looks good to merge, or you still want to add something?

shashikdm:
Yes you can merge. Please note that logger interface is not defined yet for images. Once we finalize the solution to wrapper types we can move ahead with the logger interface.

PhilipVinc:
Great! Thanks a lot!

I merged the wrapper interface an hour ago. I'm not satisfied 100% with it, so it will probably change, but for now it will suffice.  If you want you can add the relevant wrappers.

PR #22, opened 2019-05-14
Add support for Matrix and List in LogText.jl

shashikdm:
- updated the function `text_summary`
- added couple of tests


shashikdm:
s1
s2
tensorboard autmatically makes tables and lists. we only have to log the tensors with correct dimensions and values.

PhilipVinc:
Very nice!

PR #55, opened 2020-01-08
add LogEmbeddings

shashikdm:
add test for LogEmbeddings
add example for LogEmbeddings

### DIFF
--- a/examples/Embeddings.jl
+++ b/examples/Embeddings.jl
@@ -24,10 +24,5 @@ vt = svd_data.Vt[1:features, :]
 data = u*s
 metadata = MNIST.labels()[1:n]
 
-#using logger interface
-with_logger(logger) do
-    @info "embedding/mnist/loggerinterface" TBEmbeddings(data, metadata = metadata, img_labels = TBImages(imgs, HWN))
-end
-
 #using explicit function interface
 log_embeddings(logger, "embedding/mnist/explicitinterface", data, metadata = metadata, img_labels = TBImages(imgs, HWN), step = 1)
--- a/examples/Embeddings.jl
+++ b/examples/Embeddings.jl
@@ -2,18 +2,27 @@ using TensorBoardLogger #import the TensorBoardLogger package
 using Logging #import Logging package
 using TestImages
 using Flux, Flux.Data.MNIST
+using LinearAlgebra
 
 logger = TBLogger("embeddinglogs", tb_append) #create tensorboard logger
 
 ################log embeddings example: MNIST data################
-mat = MNIST.images()[1:100]
-n = size(mat, 1)
-mat = hcat(mat...)
-mat = reshape(mat, (28, 28, n))
-imgs = mat
-mat = permutedims(mat, (3, 1, 2))
-mat = reshape(mat, n, 28*28)
+n_samples = 100
+features = 100
+data = MNIST.images()[1:n_samples]
+n = size(data, 1)
+data = hcat(data...)
+data = reshape(data, (28, 28, n))
+imgs = data
+data = permutedims(data, (3, 1, 2))
+data = reshape(data, n, 28*28)
+data = convert(Array{Float64,2}, data)
+svd_data = svd(data)
+u = svd_data.U
+s = Array(Diagonal(svd_data.S))[:, 1:features]
+vt = svd_data.Vt[1:features, :]
+data = u*s
 metadata = MNIST.labels()[1:n]
 
 #using explicit function interface
-log_embeddings(logger, "mnist", mat, metadata = metadata, img_labels = TBImages(imgs, HWN), step = 0)
+log_embeddings(logger, "mnist", data, metadata = metadata, img_labels = TBImages(imgs, HWN), step = 0)
--- a/examples/Embeddings.jl
+++ b/examples/Embeddings.jl
@@ -7,11 +7,13 @@ logger = TBLogger("embeddinglogs", tb_append) #create tensorboard logger
 
 ################log embeddings example: MNIST data################
 mat = MNIST.images()[1:100]
+n = size(mat, 1)
 mat = hcat(mat...)
-mat = reshape(mat, (28, 28, 100))
+mat = reshape(mat, (28, 28, n))
+imgs = mat
 mat = permutedims(mat, (3, 1, 2))
-mat = reshape(mat, 100, 28*28)
-metadata = MNIST.labels()[1:100]
+mat = reshape(mat, n, 28*28)
+metadata = MNIST.labels()[1:n]
 
 #using explicit function interface
-log_embeddings(logger, "mnist", mat, metadata = metadata, step = 0)
+log_embeddings(logger, "mnist", mat, metadata = metadata, img_labels = TBImages(imgs, HWN), step = 0)
--- a/examples/Embeddings.jl
+++ b/examples/Embeddings.jl
@@ -30,4 +30,4 @@ with_logger(logger) do
 end
 
 #using explicit function interface
-log_embeddings(logger, "embedding/mnist/explicitinterface", data, metadata = metadata, img_labels = TBImages(imgs, HWN), step = 0)
+log_embeddings(logger, "embedding/mnist/explicitinterface", data, metadata = metadata, img_labels = TBImages(imgs, HWN), step = 1)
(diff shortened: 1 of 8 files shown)


shashikdm:
## Todo: 
1. support image labels for points
2. implement logger interface
## Edit:
1. ~~support image labels for points~~
2. implement logger interface
3. resolve failing tests
## Edit:
1. ~~support image labels for points~~
2. ~~implement logger interface~~
3. ~~resolve failing tests~~

src/Loggers/LogEmbeddings.jl
```diff
@@ -0,0 +1,58 @@
...
+
+function write_matrix(mat::AbstractArray, matrix_path::AbstractString)
+    matrix_path = joinpath(matrix_path, "tensor.tsv")
+    file = open(matrix_path, "w")
+    mat = convert(Array{Float64,2}, mat)
+    for i in 1:size(mat, 1)
+        write(file, join(mat[i, :], '\t'))
+        write(file, '\n')
+    end
+    flush(file)
```
oxinabox: You don't need `flush` if you are going to `close` it right after.

`close`  calls `flush`

src/Loggers/LogEmbeddings.jl
```diff
@@ -0,0 +1,58 @@
...
+        end
+        write_metadata(metadata, matrix_path, metadata_header)
+    end
+    write_matrix(mat, matrix_path)
+    write_pbtext(name, logger.logdir, matrix_path, metadata, step)
+end
+
+function write_matrix(mat::AbstractArray, matrix_path::AbstractString)
+    matrix_path = joinpath(matrix_path, "tensor.tsv")
+    file = open(matrix_path, "w")
```
oxinabox: Rather than using `open`-`close` pairs.
I suggest using:
```
open(matrix_path, "w") do file
...
end
```

which will automatically call `close` at th e end of the block.
and also handle closing in case of errors being thrown etc.

src/Loggers/LogEmbeddings.jl
```diff
@@ -0,0 +1,58 @@
+"""
+
+"""
+function log_embeddings(logger::TBLogger, name::AbstractString, mat::AbstractArray; metadata=nothing, labels=nothing, metadata_header=nothing, step=nothing)
+    @assert ndims(mat) == 2 "Embedding matrix must be 2-D"
```
oxinabox: Why not dispatch on `AbstractMatrix` rather than `AbstractArray`?

src/Loggers/LogEmbeddings.jl
```diff
@@ -0,0 +1,58 @@
...
+    matrix_path = joinpath(logger.logdir, string(step), name)
+    mkpath(matrix_path)
+    if metadata != nothing
+        @assert length(metadata) == size(mat, 1) "#labels must be equal to #samples"
```
oxinabox: Why are we using an assertion rather than throwning an error?

src/Loggers/LogEmbeddings.jl
```diff
@@ -0,0 +1,58 @@
...
+        if metadata_header == nothing
+            metadata_header = [string(x) for x in metadata]
+        else
+            @assert length(metadata_header) == size(metadata, 2) "length of header must be equal to the number of columns in metadata"
+            metadata = vcat([join(metadata_header, '\t')], [join(x, '\t') for x in metadata])
```
oxinabox: Do you think the following might be clearer:
```suggestion
            metadata = [join(metadata_header, '\t'); join(x, '\t') for x in metadata])
```

examples/Embeddings.jl
```diff
@@ -0,0 +1,17 @@
+using TensorBoardLogger #import the TensorBoardLogger package
+using Logging #import Logging package
```
oxinabox: What are these comments for?

examples/Embeddings.jl
```diff
@@ -0,0 +1,17 @@
...
+logger = TBLogger("embeddinglogs", tb_append) #create tensorboard logger
+
+################log embeddings example: MNIST data################
+mat = MNIST.images()[1:100]
+mat = hcat(mat...)
+mat = reshape(mat, (28, 28, 100))
+mat = permutedims(mat, (3, 1, 2))
+mat = reshape(mat, 100, 28*28)
+metadata = MNIST.labels()[1:100]
+
```
oxinabox, line 26: THis doesn't seems like a good thing to be embedding.
I suggest either:
 - Use word embeddings from Embeddings.jl
 - Use SVD or similar to find a low dimensional representation of MNIST.

shashikdm: used SVD in the example

shashikdm:
@oxinabox @PhilipVinc How should we achieve logger dispatch for embeddings? it's inputs are 2-D data matrix, optional metadata, metadata_header, img_labels. Perhaps a wrapper: `TBEmbeddings`?

test/runtests.jl
```diff
@@ -212,7 +212,6 @@ end
             @info "test3" i=i j=2^i dd=rand(10).-0.1*i hh=data_tuple log_step_increment=0
         end
     end
-
     @test TensorBoardLogger.step(logger) == 10
```
shashikdm, line 216: @PhilipVinc @oxinabox  this test fails in some runs and passes in others in Travis CI. If I run this snippet in Julia REPL, it passes. what could cause such strange behavior? 🤔

PhilipVinc:
yes, for now I would declare a wrapper.

shashikdm:
The issue with embedding is that it does not write anything to the main file (events.out...) of tensorboard. it creates other files (projector_config.pbtext, tensor.tsv, metadata.tsv, sprite.png). the file project_config.pbtext must be in the same folder as main file. so `log_embedding` requires the log directory to write the files. this means there will not be a `summary_impl` function for embeddings and `preprocess` function will directly call `log_embedding`. The only problem is, through `preprocess`, we cannot send the log directory. I have a suggestion to make `TBLogger` class singleton. that way, we can write the `logdir` function without specifying the `TBLogger` object. @PhilipVinc @oxinabox

PhilipVinc:
Making `TBLogger` a singleton is a bad idea, going contrary to the fact that we have logging contexts / multiple loggers might exist.

shashikdm:
yes that's right. my bad 😅

shashikdm:
how about we pass the logger object to all `preprocess` functions. Doing so will preserve the multiple dispatch setting and will also satisfy `log_embedding`'s requirement. I doubt if it will have much effect on performance 🤔
Edit:
I'm trying to do something like:
```
preprocess(lg::TBLogger, name, x::TBEmbedding, data) = log_embeddings(lg, name, x.mat, x.metadata, x.metadata_header, x.img_labels)
preprocess(lg::TBLogger, name, x, data) = preprocess(name, x, data)
```
Then in `handle_message` function:
```
preprocess(message*"/$key", val, data) -> preprocess(lg, message*"/$key", val, data)
```

shashikdm (commented):
From playing around in the other branch, I found out that the cause is `testimages`. when it has to download the images (ie for the first time), it causes `handle_message` to be called twice.

test/runtests.jl
oxinabox: I don't know. were their other failures?

shashikdm: figured it out. it was happening whenever `testimages` downloaded data inside the do block, it would somehow call `handle_message` function and cause step to increment

PhilipVinc:
Sorry for the delay. I'm overwhelmed by work those days...

> how about we pass the logger object to all preprocess functions.

I'm not sure I like this idea. Of course it does not change much, but as of now I feel the logic is quite clean (the code to serialise log events is decoupled to I/O and having a logger by itself). Passing the logger too just for a special case seems unnecessary.

Moreover, I feel `LogEmbeddings`, as `LogGraph` is somewhat a _special_ case of log message, which happens only once per session (please correct me if I'm wrong), so going down the same path as other more standard messages seems a bit of a stretch to me.

shashikdm:
yeah it makes sense for `log_embeddings` to be called only once unlike `log_value`. in any case, in order to write an embedding to tensorboard, the log directory is absolutely necessary. if we can't pass the logdir to `preprocess` then should we not give a logger interface for embeddings?

PhilipVinc:
I'd prefer if in this PR we just merge the back-end support for embeddings and the explicit interface.
For the log interface, I'd like to think a bit more about it.

PhilipVinc:
LGTM. @oxinabox let's merge this?

PR #91, opened 2021-03-12
Add a read-only initpolicy

Eggiverse:
When I try to load an event file and covert it to `MVHistory`, I have to use `tb_append` which creates a new event file
which is quite annoying.

I can avoid this by using `map_summaries` of course, but this is not as convenient as direct conversion.

It would be nice to have a read-only `InitPolicy`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TensorBoardLogger"
 uuid = "899adc3e-224a-11e9-021f-63837185c80f"
 authors = ["Filippo Vicentini <filippovicentini@gmail.com>"]
-version = "0.1.14"
+version = "0.1.15"
 
 [deps]
 CRC32c = "8bf52ea8-c179-5cab-976a-9e18b702a9bc"
--- a/src/TBLogger.jl
+++ b/src/TBLogger.jl
@@ -1,12 +1,11 @@
 mutable struct TBLogger <: AbstractLogger
     logdir::String
-    file::IOStream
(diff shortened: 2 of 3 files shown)


PhilipVinc:
Hi, sorry for the delay.

I will be able to devote time to this on Sunday.
If you can ping me again on Sunday, this would help me remember. I aim super-busy those days.

In any case, I am not a fan of adding an init policy like that, because TBLogger is not really a file handle. 
But maybe I am wrong. I dunno if @oxinabox might have an opinion , but rather think we should add another function to deserialise arbitrary objects fed at it.

oxinabox:
I think i agree, we should have a seperate function for reading things.
It doesn't need most of the functionality of the logger (though can share some methods for implementation)

Eggiverse:
Maybe we should have a type that does not inherit `AbstractLogger` but  works like a `TBLogger` for deserialisation only.

```julia
struct TBLogReader
...
end

function Base.convert(::Type{MVHistory}, tbl::TBLogReader; kwargs...)
...
end
```

PhilipVinc:
Yes, that would be much better.

--
Filippo Vicentini
Collaborateur Scientific
EPFL, Lausanne
Google Scholar
+41 21 693 58 38
Il 25 mar 2021, 09:34 +0100, Zhang Shitian ***@***.***>, ha scritto:

Eggiverse:
I am closing this with new pull request #94

PR #119, opened 2023-03-01
Fix MultilineChartContent not defined

ldeso:
Calling the function [log_custom_scalar](https://julialogging.github.io/TensorBoardLogger.jl/dev/explicit_interface/#TensorBoardLogger.log_custom_scalar) results in an error:
```
ERROR: UndefVarError: `MultilineChartContent` not defined
```
This fixes the error by importing the missing undefined classes.

Those classes are defined in [src/protojl/tensorboard/plugins/custom_scalar/tensorboard/layout_pb.jl](https://github.com/JuliaLogging/TensorBoardLogger.jl/blob/e985f09/src/protojl/tensorboard/plugins/custom_scalar/tensorboard/layout_pb.jl).

This is similar to what was done in the file for the audio logger [src/Loggers/LogAudio.jl](https://github.com/JuliaLogging/TensorBoardLogger.jl/blob/e985f09/src/Loggers/LogAudio.jl#L1).

### DIFF
--- a/src/Loggers/LogCustomScalar.jl
+++ b/src/Loggers/LogCustomScalar.jl
@@ -28,14 +28,12 @@ function chart(name::String, metadata::Tuple{tb_chart_type, AbstractArray})
     chart_type, tags = metadata
 
     if chart_type == tb_multiline
-        content = MultilineChartContent(tag = tags)
-        return Chart(title = name, multiline = content)
+        content = MultilineChartContent(tags)
+        return Chart(name, OneOf(:multiline, content))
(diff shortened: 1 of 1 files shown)


ldeso:
Due to the ProtoBuf update, there are actually more changes required. Marking this as draft for now.

Edit: done I think!

ldeso:
For the record, even with the changes in this pull request there is still an issue an with the custom_scalar plugin and `map_summaries`, and I haven't managed to solve it.

PR #127, opened 2023-07-03
add native support for Gadfly plots

nomadbl:
Add native Gadfly plot support.
Includes a new usage example and docstrings.


oxinabox (approved):
We really should move all these over to Julia 1.9's extension modules / weak deps
Those can work in 1.8 too, so we would only have to restrict compat to  1.8+

I will grant you merge rights shortly so feel free to merge this when happy

nomadbl:
Merging this. There is an unfortunate annoying print to stdout when the plots are displayed, but this is a Gadfly/Compose issue (see https://github.com/GiovineItalia/Gadfly.jl/issues/1621 and https://github.com/GiovineItalia/Compose.jl/issues/440)

Issue #73, opened 2020-08-07
do-notation API for logging everything in one step


ToucheSir:
Following #8 and #9, would it be feasible to integrate an API along the lines of:
```julia
...
with_logger(lg) do
    for i=1:100
        # increments log_step_increment by one by default
        # could also take a custom increment as an optional parameter
        with_step() do
            @info "test" i=i        
        end
    end
end
```
For use-cases where one only wants to increment the step once per iteration (e.g. to match tensorflow's behaviour), this would eliminate the need to specify `log_step_increment=0` on the remaining N-1 logging calls.

nomadbl:
@ToucheSir 
Great idea!
I've implemented it in the above PR.
Would be great if you can review it.
(Shameless plug - please review my other PR's :pray: )

ToucheSir:
I'm not a maintainer of this repo and I haven't used this package for a while, but will look at the PR. A better bet for you will be to get some contributor attention instead

PR #128, opened 2023-07-05
do-notation API for logging everything in one step

nomadbl:
This implements the idea from issue #73 
This adds the helper function, docs and example code

### DIFF
--- a/src/TBLogger.jl
+++ b/src/TBLogger.jl
@@ -298,3 +298,53 @@ Base.show(io::IO, mime::MIME"text/plain", tbl::TBLogger) = begin
     """
     Base.print(io, str)
 end
+
+"""
+`with_TBLogger_hold_step(f, [step]; step_at_end::Bool=true)`
+Context function to ease control of logging steps and synchronization.
+Amount of step increment can be controlled via `set_step_increment!``.
+
+Example:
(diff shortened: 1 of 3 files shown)


src/TBLogger.jl
```diff
@@ -298,3 +298,53 @@ Base.show(io::IO, mime::MIME"text/plain", tbl::TBLogger) = begin
...
+    set_step!(logger, step)
+    set_step_increment!(logger, 0)
+    f()
+    set_step!(logger, curr_step)
+    set_step_increment!(logger, curr_increment)
+    step_at_end && increment_step!(logger, curr_increment)
+end
+function with_TBLogger_hold_step(f; step_at_end::Bool=true)
+    logger = CoreLogging.current_logger()
+    @assert logger isa TBLogger "with_TBLogger_hold_step: current logger is not a TBLogger, cannot establish current step automatically"
```
oxinabox: this should be:
```suggestion
    isa(logger, TBLogger) || error("with_TBLogger_hold_step: current logger is not a TBLogger, cannot establish current step automatically")
```
as assertions may be removed depending how optiions are set

nomadbl:
Thanks, didn't know that about asserts!

PR #129, opened 2023-07-13
Add hparams API

JamieMair:
Continues work from https://github.com/JuliaLogging/TensorBoardLogger.jl/issues/77. 

EDIT: This now works! Closes https://github.com/JuliaLogging/TensorBoardLogger.jl/pull/87


oxinabox:
I am excited about this feature.

.gitignore
```diff
@@ -7,3 +7,5 @@ docs/Manifest.toml
 
 gen/proto
 gen/protojl
+
+.vscode/
```
nomadbl: In my understanding we shouldn't add IDE specific things here. I recommend using a global `.gitignore` like described here https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/
This will also save you work in future projects.
```suggestion
```

src/hparams.jl
```diff
@@ -0,0 +1,126 @@
+import .tensorboard_plugin_hparams.hparams: Interval, MetricInfo, MetricName, HParamInfo, Experiment, HParams
+import .tensorboard_plugin_hparams.hparams: var"#DataType" as HParamDataType, DatasetType as HDatasetType
+import .tensorboard_plugin_hparams.google.protobuf: ListValue as HListValue, OneOf as HOneOf, Value as HValue
```
nomadbl: Why are `OneOf`, `ListValue` etc imported here instead of `src/TensorBoardLogger.jl`?
And why the aliases?
For example OneOf is already imported and should be usable.

src/hparams.jl
```diff
@@ -0,0 +1,126 @@
...
+
+    for s in (experiment_summary, session_start_summary, session_end_summary)
+        event = make_event(logger, s)
+        write_event(logger, event)
+    end
+    nothing
+end
+
+
+export write_hparams!
```
nomadbl: move this export to `src/TensorBoardLogger.jl`

src/hparams.jl
```diff
@@ -0,0 +1,126 @@
...
+    domain::Union{Nothing, HParamRealDomain, HParamSetDomain} = nothing
+end
+Base.@kwdef struct MetricConfig
+    name::String
+    displayname::String = ""
+    description::String = ""
+end
+
+default_domain(::Val{Bool}) = HParamSetDomain([false, true])
+default_domain(::Val{Float64}) = HParamRealDomain(typemin(Float64), typemax(Float64))
```
nomadbl, line 27: Shouldn't this be more flexible?
```suggestion
default_domain(f::Val{AbstractFloat}) = HParamRealDomain(typemin(f), typemax(f))
```

src/hparams.jl
```diff
@@ -0,0 +1,126 @@
...
+    PLUGIN_NAME = "hparams"
+    PLUGIN_DATA_VERSION = 0
+
+    EXPERIMENT_TAG = "_hparams_/experiment"
+    SESSION_START_INFO_TAG = "_hparams_/session_start_info"
+    SESSION_END_INFO_TAG = "_hparams_/session_end_info"
+
+    # Check for datatypes
+    for v in values(hparams)
+        @assert typeof(v) <: Union{Bool, String, Float64} "Hyperparameters must be of types String, Bool or Float64"
```
nomadbl: Alternatively, we could promote any AbstractFloat here into Float64 if supporting non Float64 is a problem.

nomadbl (commented):
Seems good to me, other than the fact that it doesn't register more than 1 hparam.
I haven't been able to track down the problem yet.
I would change what I indicated, especially the Float64 issue so users do not have to change data types explicitly.

.gitignore
JamieMair: Done - this workaround is interesting - thanks!

src/hparams.jl
JamieMair: The `ListValue` is specific to the hparams protobuf library, but I have used the original `OneOf`. I have also removed most of the imports as to not clutter the namespace, but HListValue is still there. I will move these imports before going out of draft.

src/hparams.jl
JamieMair: Done :)

src/hparams.jl
JamieMair: It seems that the hparams plugin only supports `String`, `Bool` or `Float64`. I have took this suggestion and made the API allow any `Real` number which is then cast to a `Float64` and explained this in the docstring.

src/hparams.jl
JamieMair: As previous comment - allows any Real but promotes.

JamieMair:
@oxinabox @nomadbl I have finished the feature. The way that dictionary was serialised was actually wrong in the protobuf code, so I made it match the output of tensorboardX now. Thankfully, this fixed the issue of only one parameter showing up.

I have added some docs as well on how to use it, along with an example script. Should be simple enough as users should only really interact with `write_hparams!`. 

Note, I also bumped the version of the package as well. Should be no breaking changes as this only adds functionality.

nomadbl:
There's some really great work here, kudos.
However it looks like this points to a bug in `Protobuf.jl`.
The functions which are edited here already exist in https://github.com/JuliaIO/ProtoBuf.jl/blob/master/src/codec/encode.jl
Please file an issue there, as your understanding of the inner workings there is better than mine.
We should probably wait until we resolve it there and then we wouldn't need to add functions on our end which may break in the future if any changes to Protobuf.jl are made.

JamieMair:
> There's some really great work here, kudos. 

Thanks! 

> However it looks like this points to a bug in `Protobuf.jl`. 

I have opened an issue (https://github.com/JuliaIO/ProtoBuf.jl/issues/233) and PR (https://github.com/JuliaIO/ProtoBuf.jl/pull/234), but these would be breaking changes for any downstream users, so I think that there is a chance that they might not accept the changes - it may take a while for this to go anywhere.

> We should probably wait until we resolve it there and then we wouldn't need to add functions on our end which may break in the future if any changes to Protobuf.jl are made.

As the added methods have strict types, only applying to the HParams plugin, how would you feel about merging these changes now, as any changes in ProtoBuf may take a while/not happen? I have added a unit test with what the exact serialisation and deserialisation should be, which will make it easier to transition to any upstream fixes while verifying that tensorboard can still parse the output.

test/test_hparams.jl
```diff
@@ -34,4 +35,46 @@ test_hparams_log_dir = "test_hparams_logs/"
...
+
+
+    hparams_dict = Dict(k => TensorBoardLogger._convert_value(v) for (k, v) in hparams_config)
+
+    session_start_info = TensorBoardLogger.HP.SessionStartInfo(hparams_dict, "", "", "", zero(Float64))
+    session_start_content = TensorBoardLogger.HP.HParamsPluginData(PLUGIN_DATA_VERSION, TensorBoardLogger.OneOf(:session_start_info, session_start_info))
+
+    expected_bytes = UInt8[0x1a, 0x5b, 0x0a, 0x0e, 0x0a, 0x06, 0x6d, 0x65, 0x74, 0x68, 0x6f, 0x64, 0x12, 0x04, 0x1a, 0x02, 0x4d, 0x43, 0x0a, 0x12, 0x0a, 0x05, 0x61, 0x6c, 0x70, 0x68, 0x61, 0x12, 0x09, 0x11, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xe0, 0x3f, 0x0a, 0x0f, 0x0a, 0x02, 0x69, 0x64, 0x12, 0x09, 0x11, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xf0, 0x3f, 0x0a, 0x13, 0x0a, 0x0d, 0x6f, 0x70, 0x74, 0x69, 0x6d, 0x69, 0x73, 0x61, 0x74, 0x69, 0x6f, 0x6e, 0x73, 0x12, 0x02, 0x20, 0x00, 0x0a, 0x0f, 0x0a, 0x02, 0x70, 0x31, 0x12, 0x09, 0x11, 0x9a, 0x99, 0x99, 0x99, 0x99, 0x99, 0xb9, 0x3f]
+    
+    @test TensorBoardLogger.encode_bytes(session_start_content) == expected_bytes
```
nomadbl: Great idea.
Lets also put this test under a seperate "testset" with an appropriate title, so it is easy to know what to do if it fails.

nomadbl (approved):
I think with the precautions you have added it is reasonable to merge this now.
Thanks for taking care of the Protobuf issue and PR as well!

I have added another suggestion and approved the PR. Merge as you see fit.

test/test_hparams.jl
JamieMair: Agreed, I have separated them into two test cases now.

JamieMair:
@nomadbl Thanks for the quick review and guidance. I'll updated from master and will merge as soon as the unit tests pass.

JamieMair:
@nomadbl I have realised that I don't actually have write access, I think someone else would need to merge the changes. Thanks :)

PR #132, opened 2023-07-21
Fix serialisation bug

### LINKED ISSUE #131: Error when reading tfevents files written with TensorBoardLogger v0.1.21
Hello, I updated `TensorboardLogger` to the latest `0.1.21` version, and the following minimum example doesn't work anymore:

```julia
using Logging
using TensorBoardLogger
using ValueHistories

logger = TBLogger("logs"; min_level=Logging.Info)

with_logger(logger) do
    @info "train" loss = 1.0
    @info "train" loss = 0.0
end

r = TBReader("logs1")
hist = MVHistory()

TensorBoardLogger.map_summaries(r) do tag, iter, val
    push!(hist, Symbol(tag), iter, val)
end
```

Here is the error:

```bash
ERROR: AssertionError: crc_data == crc_data_ck
Stacktrace:
 [1] read_event(f::IOStream)
   @ TensorBoardLogger ~/.julia/packages/TensorBoardLogger/LO5eP/src/Deserialization/deserialization.jl:49
 [2] iterate(it::TensorBoardLogger.TBEventFileIterator, state::Int64)
   @ TensorBoardLogger ~/.julia/packages/TensorBoardLogger/LO5eP/src/Deserialization/deserialization.jl:126
 [3] map_summaries(fun::var"#21#22", logdir::TBReader; purge::Bool, tags::Nothing, steps::Nothing, smart::Bool)
   @ TensorBoardLogger ~/.julia/packages/TensorBoardLogger/LO5eP/src/Deserialization/deserialization.jl:270
 [4] map_summaries(fun::Function, logdir::TBReader)
   @ TensorBoardLogger ~/.julia/packages/TensorBoardLogger/LO5eP/src/Deserialization/deserialization.jl:248
```

Reading files written with an older version still works.
Am I missing something ?

ps: by the way, it seems that version `0.1.20` cannot be installed from the general registry for some reason (only from the github repository url)

JamieMair:
Closes https://github.com/JuliaLogging/TensorBoardLogger.jl/issues/131

See commits for explanation.

@nomadbl would you be able to review & merge this fix?

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -9,7 +9,7 @@ end
 function serialize_proto(data)
     pb = PipeBuffer()
     encode(ProtoEncoder(pb), data)
-    return view(pb.data, 1:pb.size) # Do not return additional zeros
+    return take!(pb) # Do not return additional zeros
 end
 
 """
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -9,7 +9,7 @@ end
 function serialize_proto(data)
     pb = PipeBuffer()
     encode(ProtoEncoder(pb), data)
-    pb.data
+    return view(pb.data, 1:pb.size) # Do not return additional zeros
 end
 
 """
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TensorBoardLogger"
 uuid = "899adc3e-224a-11e9-021f-63837185c80f"
 authors = ["Filippo Vicentini <filippovicentini@gmail.com>"]
-version = "0.1.21"
+version = "0.1.22"
 
 [deps]
 CRC32c = "8bf52ea8-c179-5cab-976a-9e18b702a9bc"
--- a/src/event.jl
+++ b/src/event.jl
@@ -29,17 +29,16 @@ format. The format follows the following rule (in bytes)
 #4 N..N+8      UInt32  masked_CRC of #3
 """
 function write_event(out::IO, event::Event)
-    data = PipeBuffer();
-    encode(ProtoEncoder(data), event)
+    event_bytes = serialize_proto(event)
 
     #header
-    header     = collect(reinterpret(UInt8, [data.size]))
+    header     = collect(reinterpret(UInt8, [length(event_bytes)]))
     crc_header = reinterpret(UInt8, UInt32[masked_crc32c(header)])
-    crc_data   = reinterpret(UInt8, UInt32[masked_crc32c(data.data)])
+    crc_data   = reinterpret(UInt8, UInt32[masked_crc32c(event_bytes)])
 
     write(out, header)
     write(out, crc_header)
-    write(out, data.data)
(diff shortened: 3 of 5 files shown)


src/utils.jl
```diff
@@ -9,7 +9,7 @@ end
 function serialize_proto(data)
     pb = PipeBuffer()
     encode(ProtoEncoder(pb), data)
-    pb.data
+    return view(pb.data, 1:pb.size) # Do not return additional zeros
```
nomadbl: @JamieMair 
This seems like the core of this PR (and the other changes are for reuse of this in other places).
But I can't understand the difference between this and the previous implementation.
Can you give a short explanation and links?

Overall the PR looks good I just want to do some due diligence

JamieMair: Under the hood, `ProtoBuf.jl` uses the function `maybe_ensure_room` ([link](https://github.com/JuliaIO/ProtoBuf.jl/blob/548017cb69778cb5f0cd1273b16507f81d64719e/src/codec/encode.jl#L7C2-L7C2)), which expands an IO buffer to potentially hold more bytes (padding zeros usually), so that new bytes can be written without unnecessary resizing. Sometimes this guesses the amount of bytes to expand by and overshoots (as with the example in https://github.com/JuliaLogging/TensorBoardLogger.jl/issues/131), but the padding bytes should not be written to the events file, as it causes parsing errors. 

In https://github.com/JuliaLogging/TensorBoardLogger.jl/issues/131, the size of an event wrote the size of the unpadded bytes to the file, but the when skipping past the actual payload, ran into the padded zero bytes and tried to read this as the checksum, which caused the crc error. Also the checksum should only be ran on the unpadded data.

I actually ran into this problem with the hparams implementation but didn't really understand why `serialize_proto` wouldn't work for my application, which is why I wrote the function with `take!`. Now that I know why it wasn't working, I figured I should also update the hparams implementation to use the updated `serialize_proto` function instead. The main difference between this and my old function is the `view` which avoids the extra allocation.

nomadbl: Lets change back to the `take!` version.
It looks like it does not allocate and we would avoid possible future breaking changes.

note:
I tested `take!` using
```julia
julia> using BenchmarkTools
julia> pb = PipeBuffer()
julia> write(pb, rand(1000))
julia> @benchmark take!(pb)
Time  (median):      37.376 ns
Memory estimate: 64 bytes, allocs estimate: 1.

julia> write(pb, rand(1000000))
julia> @benchmark take!(pb)
Time  (median):      36.753 ns
Memory estimate: 64 bytes, allocs estimate: 1.
```

nomadbl:
```suggestion
    return take!(pb) # Do not return additional zeros
```

nomadbl (approved):

JamieMair:
@nomadbl I have taken the suggestion, thanks!

Issue #137, opened 2023-09-18
`convert(MVHistory, logger)` fails if hyperparameters have been recorded


torfjelde:
MWE:

``` julia
julia> using TensorBoardLogger, Logging, ValueHistories

julia> lg = TBLogger(joinpath(mktempdir(), "runs"), min_level=Logging.Info; step_increment=0)
TBLogger:
    - Log level     : Info
    - Current step  : 0
    - Output        : /tmp/jl_HWNylZ/runs
    - open files    : 1


julia> TensorBoardLogger.write_hparams!(
           lg,
           Dict("hi" => 1.0),
           ["x/val"]
       )

julia> with_logger(lg) do
           @info "x" val=3.0
       end

julia> convert(MVHistory, lg)
ERROR: Summary value of Nothing while deserializing
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] summary_type
   @ ~/.julia/packages/TensorBoardLogger/pDb23/src/Deserialization/deserialization.jl:141 [inlined]
 [3] iterate(iter::TensorBoardLogger.SummaryDeserializingIterator, state::Int64)
   @ TensorBoardLogger ~/.julia/packages/TensorBoardLogger/pDb23/src/Deserialization/deserialization.jl:190
 [4] iterate
   @ ~/.julia/packages/TensorBoardLogger/pDb23/src/Deserialization/deserialization.jl:184 [inlined]
 [5] map_summaries(fun::TensorBoardLogger.var"#131#132"{MVHistory{History}}, logdir::TBLogger{String, IOStream}; purge::Bool, tags::Nothing, steps::Nothing, smart::Bool)
   @ TensorBoardLogger ~/.julia/packages/TensorBoardLogger/pDb23/src/Deserialization/deserialization.jl:253
 [6] map_summaries
   @ ~/.julia/packages/TensorBoardLogger/pDb23/src/Deserialization/deserialization.jl:237 [inlined]
 [7] #convert#130
   @ ~/.julia/packages/TensorBoardLogger/pDb23/src/Optional/ValueHistories.jl:5 [inlined]
 [8] convert(::Type{MVHistory}, tbl::TBLogger{String, IOStream})
   @ TensorBoardLogger ~/.julia/packages/TensorBoardLogger/pDb23/src/Optional/ValueHistories.jl:3 [9] top-level scope
   @ REPL[34]:1
```

In contrast, the following works:

``` julia
[...]

PR #138, opened 2023-09-18
Fix for #137

torfjelde:
Fix for #137

Not sure if this is the way to go, but figured I'd submit it in case it's useful.

### DIFF
--- a/test/test_hparams.jl
+++ b/test/test_hparams.jl
@@ -89,3 +89,18 @@ end
         @test expected_hparams_config[k] == decoded_v
     end
 end
+
+@testset "Reading with hparams present (#137)" begin
+    # https://github.com/JuliaLogging/TensorBoardLogger.jl/issues/137
+    lg = TBLogger(joinpath(mktempdir(), "logs"))
+    TensorBoardLogger.write_hparams!(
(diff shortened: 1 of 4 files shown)


test/test_hparams.jl
```diff
@@ -88,4 +88,19 @@ end
...
+    lg = TBLogger(joinpath(mktempdir(), "logs"))
+    TensorBoardLogger.write_hparams!(
+        lg,
+        Dict("hi" => 1.0),
+        ["x/val"]
+    )
+    with_logger(lg) do
+        @info "x" val=3.0
+    end
+    hist = convert(MVHistory, lg)
```
oxinabox, line 104: this is the line that is failing. Is what we need:
```suggestion
    hist = convert(ValueHistories.MVHistory, lg)
```
Or do we need to make sure ValueHistories.jl is loaded?

torfjelde: Ah I think I've gotten confused by the fact that ValueHistories.jl is listed as a dep for the `test` target in the Project.toml, and thus assumed it was already loaded. I loaded it in `runtests.jl` now:) Should hopefully do the trick.

PR #150, opened 2025-09-26
Add support for sampling rate in audio logging

### LINKED ISSUE #149: Sampling rate not correctly processed
Hi folks! Great package and really easy to use, kudo's for that!

I was just logging audio as well and noted that the durations of the audio being saved did not correspond to the duration of the audio displayed in Tensorboard. The examples in [`examples/Audios.jl`](https://github.com/JuliaLogging/TensorBoardLogger.jl/blob/master/examples/Audios.jl) seem to have a duration in Tensorboard of 11 seconds. It's like the sampling rate does not matter and that it defaults to something like 8 kHz.

bartvanerp:
Resolves #149 

TensorBoardLogger.jl currently writes WAVs and puts them in encoded_audio_string. That means TensorBoard will only pay attention to the sample rate stored in the WAV header. The sample rate is not written there however. It is not possible to include the header information in the `FileIO.stream`, hence using `wavwrite` and adding `WAV.jl` is necessary as this does support the sample rate argument.


oxinabox:
nice work.
Can we write a unit test for this?

PR #151, opened 2026-06-13
Update ProtoBuf to 1.x and regenerate protobuf definitions

KristianHolme:
> **Note:** This PR was authored with [Cursor](https://cursor.com) (AI-assisted coding).

## Summary

Fixes the ProtoBuf.jl deprecation warning:

> You are using code generated by an older version of ProtoBuf.jl, which was deprecated. Please regenerate your protobuf definitions with the current version of ProtoBuf.jl.

- **Regenerate `src/protojl/`** with ProtoBuf.jl **1.3** (new `decode(d, T, _endpos, _group)` API)
- **Bump compat:** `ProtoBuf` `1.0.11` → `1`, Julia `1.6` → `1.10` (required by ProtoBuf 1.3)
- **Source fixes** for the new codegen:
  - Restore plugin module names (`tensorboard_plugin_text`, etc.) after regeneration
  - Update `hparams.jl` for new `HParamInfo.differs` field
  - Remove obsolete custom `Dict` encode/decode workarounds (ProtoBuf 1.3 handles maps natively)
  - Drop dead `writeproto` / legacy `hasproperty` code
- **Codegen tooling:** `compile_proto.jl` now replaces existing output and renames plugin modules; `download_proto_source` fixes Linux `sed` for `struct.proto` → `struct_tb.proto`

## Test plan

- [x] `Pkg.test()` — **166 passed**, 1 failed (optional Plots PNG bytes), 1 broken (Graph Logger; pre-existing)
- [x] All hparams tests pass (encode/decode round-trip verified)
- [x] No ProtoBuf deprecated-codegen warning on load/encode

Made with [Cursor](https://cursor.com)


oxinabox:
Can you look at why CI is failing?

KristianHolme:
Looks like the failing tests was some tests assuming byte-identical output when that is not really guaranteed, and some windows file deletion issues. I'm not super-familiar with this codebase, but seems reasonable to me :)

🤖 Written by GPT-5.6-Sol:

## TL;DR

- **Plots:** Replaced raw PNG byte comparison with checks for valid PNG output and matching image attributes.
- **HParams:** Replaced order-sensitive protobuf byte comparison with a semantic encode/decode round-trip.
- **Logger cleanup:** Fixed `finalize(nothing)` by finalizing the logger directly and verifying its streams close and files can be deleted.
- **Isolation:** Temporary directories and explicit logger closure prevent Windows file-lock failures and cascading test errors.

<details><summary>Details</summary>
<p>

## CI failure analysis and test fixes

The failures came from three brittle tests rather than a Plots dependency installation failure.

### Plots test

#### Previous test

```julia
pb = PipeBuffer()
show(pb, MIME("image/png"), p)
data = TensorBoardLogger.preprocess("key", p, Vector())

@test last(data[1]).data == pb.data
```

#### Intent

This attempts to verify that preprocessing a `Plots.Plot` produces the same PNG as rendering the plot directly.

#### Problem

It renders the plot twice and compares the compressed PNG byte streams. PNG output can differ because of compression, metadata, or backend state even when both images are valid and structurally equivalent.

#### Replacement

```julia
expected_image = convert(TensorBoardLogger.PngImage, p)
data = TensorBoardLogger.preprocess("key", p, Vector())

@test last(data[1]) isa TensorBoardLogger.PngImage
@test last(data[1]).attr == expected_image.attr
@test TensorBoardLogger.PNGImage.check_valid_png(last(data[1]).data)
```

This verifies that:

- the Plots-specific preprocessing dispatch is used;
- the result is a `PngImage`;
- the output is a valid PNG;
- its dimensions, colour format, bit depth, and other PNG attributes match a direct conversion.

It avoids depending on nondeterministic encoded bytes.

### HParams serialization test

#### Previous test

```julia
@test TensorBoardLogger.serialize_proto(session_start_content) ==
      hparams_metadata_encoded_bytes
```

#### Intent

This attempts to verify that HParams metadata is encoded correctly.

#### Problem

`SessionStartInfo.hparams` is a protobuf map. Protobuf does not guarantee map-entry serialization order, so two semantically identical messages can have different byte sequences.

This is why the test passed on Julia stable but failed on Julia pre when dictionary iteration order changed.

#### Replacement

```julia
encoded_content = TensorBoardLogger.serialize_proto(session_start_content)
decoder = TensorBoardLogger.ProtoDecoder(IOBuffer(encoded_content))
decoded_content = PB.decode(
    decoder,
    TensorBoardLogger.HP.HParamsPluginData,
)
decoded_session_info = decoded_content.data.value
decoded_hparams =
    Dict(k => v.kind.value for (k, v) in decoded_session_info.hparams)

@test decoded_content.version == PLUGIN_DATA_VERSION
@test decoded_content.data.name === :session_start_info
@test decoded_hparams == hparams_config
@test decoded_session_info.model_uri == ""
@test decoded_session_info.monitor_url == ""
@test decoded_session_info.group_name == ""
@test decoded_session_info.start_time_secs == 0.0
```

This tests the actual contract: the encoded message can be decoded without losing or changing its fields.

The separate golden-byte decoding test remains, so decoding compatibility with previously generated protobuf data is still covered.

### Logger finalizer test

#### Previous test

```julia
files = keys(tbl.all_files)

tbl = nothing
Base.finalize(tbl)

@test begin
    foreach(f -> rm(joinpath(test_log_dir * "run", f)), files)
    true
end
```

#### Intent

The likely intent was to verify that the logger’s registered finalizer closes its event files when the logger is no longer referenced.

#### Problem

Assigning `nothing` only removes the local reference:

```julia
tbl = nothing
```

It does not immediately run finalizers. The logger merely becomes eligible for garbage collection, and collection may happen later.

The subsequent call:

```julia
Base.finalize(tbl)
```

therefore calls `finalize(nothing)`, not `finalize` on the former logger. It does not finalize the logger.

In addition, `keys(tbl.all_files)` returns a view backed by the logger’s dictionary. Collecting the filenames avoids retaining that view during cleanup.

#### Replacement

```julia
files = collect(keys(tbl.all_files))
streams = collect(values(tbl.all_files))

Base.finalize(tbl)
tbl = nothing

@test all(stream -> !isopen(stream), streams)
@test begin
    foreach(f -> rm(joinpath(logdir, f)), files)
    true
end
```

This deterministically invokes the finalizer registered by:

```julia
Base.finalizer(Base.close, lg)
```

and verifies both observable effects:

- every event stream is closed;
- the files can be deleted, including on Windows.

This tests the registered cleanup callback directly. Testing when the garbage collector chooses to invoke it would require forced GC and weak references, making the test less deterministic.

The test also uses a temporary directory, and the reset/event tests now explicitly close their loggers to prevent open file handles from leaking into later tests.

</p>
</details>

csvance:
Would be great if this could be merged. ProtoBuf 1.0.X isn't compatible with gRPCClient.jl (or gRPCServer.jl), so machine learning workflows cannot use them currently. I'm adding support for TensorBoardLogger.jl in ReactantNitro, but I can't use it for my own experiments since I use gRPC as part of my data pipeline.

EDIT: I think I actually misunderstood the bounds here (doh). It still would be good to regenerate the bindings.

